PR #28, opened 2026-05-04
feat(transport): add OAuth session auth and Streamable HTTP transport

ecthelion77:
## Summary

This PR adds two complementary features to enable multi-tenant gateway deployments (e.g., behind an OAuth2 reverse proxy):

### OAuth per-connection authentication (`AUTH_MODE=oauth`)
- New `createMcpServer(token)` factory creates isolated Server+GitLabApi instances per connection using the Bearer token from the `Authorization` header
- PAT mode (default) is unchanged: static token from env var
- Returns 401 if `Authorization` header is missing in OAuth mode

### Streamable HTTP transport (`USE_STREAMABLE_HTTP=true`)
- Implements the MCP Streamable HTTP spec on `POST/GET/DELETE /mcp`
- Session management via `MCP-Session-Id` header
- Can coexist with legacy SSE transport on the same server
- Adds `/healthz` endpoint for readiness/liveness probes

### Helm chart values update
- Adds `AUTH_MODE` and `USE_STREAMABLE_HTTP` config keys to `chart/values.yaml`

### Other improvements
- CORS headers extended: `Authorization`, `MCP-Session-Id`, `DELETE` method
- Architecture and usage documentation added (`docs/ARCHITECTURE.md`, `docs/USAGE.md`)

## Dependencies

> **Depends on #29** — The Helm chart (`chart/`) is introduced in #29. This PR adds two config keys to `chart/values.yaml`. Merge #29 first, then this PR applies cleanly. If merged standalone (without #29), the `chart/values.yaml` change can be dropped without affecting the source code changes.

## Environment Variables
| Variable | Values | Default | Description |
| --- | --- | --- | --- |
| `AUTH_MODE` | `pat` / `oauth` | `pat` | Authentication strategy |
| `USE_STREAMABLE_HTTP` | `true` / `false` | `false` | Enable Streamable HTTP transport |
| `USE_SSE` | `true` / `false` | `false` | Enable legacy SSE transport |

## Testing
- All 41 existing tests pass
- Build succeeds with zero errors
- Tested in production with an OAuth2 gateway (multi-user GitLab access)

## Checklist
- [x] Code follows Conventional Commits format
- [x] No version bump included
- [x] Tests pass (`npm test`)
- [x] Build passes (`npm run build`)
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,67 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-_Nothing yet. New entries land here between releases._
-
-## [0.3.2] - 2026-05-04
-
-A pure security and infrastructure release. No runtime API changes; the
-server behaves identically to 0.3.1 for its consumers.
-
-### Security
-
-- `npm audit fix` resolved all 14 transitive npm vulnerabilities reported
-  against `0.3.1` (`hono`, `@hono/node-server`, `express-rate-limit`,
-  `path-to-regexp`, `picomatch`, `postcss`, `vite`). Lockfile-only changes.
-- Added `overrides.vite: ^8.0.0` in `package.json` to permanently resolve
-  three high-severity `vite` advisories (GHSA-4w7w-66w2-5vf9,
-  GHSA-v2wj-q39q-566r, GHSA-p9ff-h696-f583) that were inherited via the
-  `vitest` dev tree. `npm audit` now reports zero vulnerabilities.
-- Enabled GitHub Dependabot security updates, secret scanning, and secret
-  scanning push protection on the repository.
-- Confirmed Private Vulnerability Reporting (PRVR) enabled — disclosure
-  channel is https://github.com/yoda-digital/mcp-gitlab-server/security/advisories/new
-
 ### Added
 
-- `.github/dependabot.yml` — weekly grouped npm + GitHub Actions + Docker
-  update PRs.
-- `.github/workflows/codeql.yml` — CodeQL static analysis (security-extended
-  + security-and-quality query packs) on push, PR, and weekly schedule.
-- `.github/CODEOWNERS` — review routing for high-impact paths.
-- `.github/PULL_REQUEST_TEMPLATE.md` — pre-merge checklist anchored to
-  `CLAUDE.md` and `ai_code_of_conduct.md`.
-- `.github/ISSUE_TEMPLATE/{bug_report,feature_request,config}.yml` — issue
-  forms scoped to the MCP / GitLab / transport / auth domain. Blank issues
-  disabled; security routes to PRVR, questions route to Discussions.
-- `CODE_OF_CONDUCT.md` — Contributor Covenant 2.1 with project-specific
-  enforcement contact.
-- Branch protection ruleset on `main` (id `15919136`): pull request required,
-  status checks gated (`build-and-test` + `Analyze (javascript-typescript)`),
-  force-push and deletion blocked, linear history required, squash/rebase
-  merges only. Repository admins retain bypass for emergency hotfixes.
-- GitHub release tags `v0.3.0` and `v0.3.1` for previously published
-  versions (the `npm` registry already had them; the GitHub release timeline
-  was empty).
-
-### Changed
-
-- `.github/workflows/publish.yml`:
-  - `actions/checkout@v3` → `@v6`
-  - `actions/setup-node@v3` → `@v6`
-  - `node-version` `20.x` → `22.x` (LTS Iron)
-  - Enabled `npm test` in the `build-and-test` job (vitest is wired)
-  - `npm publish --provenance --access public` — Sigstore-signed npm
-    provenance attestations via GitHub OIDC
-  - Least-privilege `permissions:` blocks at workflow + job level
-- `CLAUDE.md`: corrected the stale "npm test exits with error" note (vitest
-  is wired); added a Security paragraph pointing at PRVR and `SECURITY.md`.
-
-### Removed
-
-- `docs/VISION.md` — superseded; product strategy is tracked elsewhere.
-- Wiki page `Product-Vision-&-Roadmap.md` — was a duplicate of
-  `docs/VISION.md`.
+- **OAuth per-connection authentication** (`AUTH_MODE=oauth`): new
+  `createMcpServer(token)` factory creates isolated Server + GitLabApi
+  instances per connection using the Bearer token from the `Authorization`
+  header. PAT mode (default) is unchanged.
+- **Streamable HTTP transport** (`USE_STREAMABLE_HTTP=true`): implements
+  the MCP Streamable HTTP spec on `POST/GET/DELETE /mcp` with session
+  management via `MCP-Session-Id` header.
+- **CORS origin allowlist** (`CORS_ALLOW_ORIGINS`): restrict allowed
+  origins in OAuth mode; permissive `*` default retained for PAT mode only.
+- `/healthz` endpoint with active session count and configurable
+  threshold (`HEALTHZ_MAX_SESSIONS`) for meaningful Kubernetes probes.
+- `docs/ARCHITECTURE.md` and `docs/USAGE.md` documentation.
 
 ## [0.3.1] - 2026-05-02
 
--- a/docs/OPERATIONS.md
+++ b/docs/OPERATIONS.md
@@ -1,0 +1,57 @@
+# Operations Guide
+
+## Health Checks
+
+The `/healthz` endpoint returns:
+
+- `200 OK` with `{"status":"ok","sessions":<n>}` when healthy.
+- `503 Service Unavailable` with `{"status":"unhealthy","reason":"session_limit_exceeded","sessions":<n>}` when active sessions exceed `HEALTHZ_MAX_SESSIONS` (default: 10000).
+
+### Kubernetes Probe Configuration
+
+```yaml
+livenessProbe:
+  httpGet:
+    path: /healthz
+    port: 3000
+  initialDelaySeconds: 5
+  periodSeconds: 30
+readinessProbe:
+  httpGet:
+    path: /healthz
+    port: 3000
+  initialDelaySeconds: 3
+  periodSeconds: 10
+```
+
+## Environment Variables
+
+| Variable | Default | Description |
+| --- | --- | --- |
+| `PORT` | `3000` | HTTP listen port |
+| `USE_SSE` | `false` | Enable legacy SSE transport |
+| `USE_STREAMABLE_HTTP` | `false` | Enable MCP Streamable HTTP transport |
+| `AUTH_MODE` | `pat` | `pat` or `oauth` |
+| `GITLAB_API_URL` | `https://gitlab.com/api/v4` | GitLab API base URL |
+| `GITLAB_PERSONAL_ACCESS_TOKEN` | — | Required in `pat` mode |
+| `GITLAB_READ_ONLY_MODE` | `false` | Restrict to read-only tools |
+| `CORS_ALLOW_ORIGINS` | — | Comma-separated allowed origins (empty = `*` in PAT mode, deny in OAuth mode) |
+| `HEALTHZ_MAX_SESSIONS` | `10000` | Session count threshold for unhealthy status |
+
+## Troubleshooting
+
+### Server returns 401 on every request
+
+- **OAuth mode**: Verify the upstream gateway is forwarding the `Authorization: Bearer <token>` header.
+- **PAT mode**: Ensure `GITLAB_PERSONAL_ACCESS_TOKEN` is set and the token has not expired.
+
+### High session count / memory growth
+
+Active sessions are stored in-memory. If session count grows unbounded:
+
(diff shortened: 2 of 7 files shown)


nalyk:
Hey @ecthelion77 — thanks for this. OAuth per-session and Streamable HTTP are both things I want in the project, and the factory shape is the right call.

Four must-fix items before I can merge:

**1. `CHANGELOG.md` — please revert.** The diff overwrites the real `0.3.1` entry (shipped 2026-05-02: `list_issues` pagination fix #24, `GITLAB_URL` doc correction, ajv/hono/rollup security patches, `.mcp.json` gitignore) and replaces it with backdated text for unreleased work. The "Listed at" footer is gone too. New work goes under `## [Unreleased]` — please restore `CHANGELOG.md` from `origin/main` and append from there.

**2. Version bump.** This is `feat()`, so `package.json` → `0.4.0`. Push-to-main auto-publishes; without a bump the workflow stalls.

**3. `chart/values.yaml`.** No `chart/` directory exists in this repo yet. This belongs in the Helm chart PR you mentioned — let's keep it out of this one.

**4. Tests for the new code paths.** Vitest is already wired (`npm test`). For something that touches per-tenant auth I'd want at least: bearer extraction (missing / malformed / case-variant → 401), Streamable HTTP session lifecycle (init → reuse → cleanup), and the cross-protocol session-collision branch. The "41 tests pass" line in the description is true, but none of those tests exercise the new code.

A couple of things I'd like addressed in this PR or as fast-follow, not blockers:

- `AUTH_MODE=Oauth` / typo silently degrades to PAT mode. Fail-fast on parse would be safer for a multi-tenant deployment.
- `/healthz` always returns `200`, including when the transports map is leaking or the process is wedged. A map-size threshold or short-cached GitLab reachability check would make Kubernetes probes meaningful.
- `docs/USAGE.md` and `docs/ARCHITECTURE.md` both reference `docs/OPERATIONS.md`, which isn't in the diff — broken cross-refs on day one.

I'll leave three more focused technical comments below on CORS, a memory-leak path on `connect()` failure, and a refactor suggestion for `createMcpServer`. None of those three are blockers on their own, but the CORS one is worth resolving before this is exposed publicly.

Happy to take this in two PRs if it's easier (factory + bearer on existing SSE first, then Streamable HTTP separately). Up to you.

Thanks again — this is real work and I appreciate it.

nalyk:
**CORS configuration interacts badly with the new OAuth mode** — `src/transport.ts` ~lines 52-54:

```ts
res.setHeader('Access-Control-Allow-Origin', '*');
res.setHeader('Access-Control-Allow-Methods', 'GET, POST, DELETE');
res.setHeader('Access-Control-Allow-Headers', 'Content-Type, Authorization, MCP-Session-Id');
```

Per the Fetch spec, `Access-Control-Allow-Origin: *` is incompatible with `Access-Control-Allow-Credentials: true` for cookies and HTTP basic auth — but an `Authorization: Bearer <token>` header set programmatically by JS is **not** a "credential" in the CORS sense. It's a custom header, and the PR explicitly allows it via `Allow-Headers`, so the preflight succeeds and any origin can send a Bearer-bearing request to `/mcp`, `/sse`, or `/messages` and read the response.

Concrete failure mode: any time a victim's bearer token leaks through a separate vector (XSS in another tool sharing the gateway origin, browser extension reading localStorage, a token accidentally pasted into a logging session), an attacker page on any domain can use that token directly via `fetch('/mcp', { headers: { Authorization: 'Bearer <stolen>' } })`. The browser will not reject the response — `Allow-Origin: *` invites it.

For an `AUTH_MODE=oauth` deployment behind a multi-tenant gateway, this undoes the isolation the new auth path is meant to provide. The PAT-mode default is more defensible (single tenant, single token, often local) — the issue is specifically the combination of wide-open CORS with delegated bearer tokens.

Suggested fix — allowlist origins explicitly via env var, with a permissive fallback only for PAT mode:

```ts
const allowedOrigins = (process.env.CORS_ALLOW_ORIGINS ?? '')
  .split(',').map(s => s.trim()).filter(Boolean);
const origin = req.headers.origin;

if (origin && allowedOrigins.includes(origin)) {
  res.setHeader('Access-Control-Allow-Origin', origin);
  res.setHeader('Vary', 'Origin');
} else if (allowedOrigins.length === 0 && AUTH_MODE === 'pat') {
  // single-tenant local dev — keep permissive default
  res.setHeader('Access-Control-Allow-Origin', '*');
}
// else: no Allow-Origin header → browser refuses to surface the response
```

If the intended use case is server-to-server only (no browser clients ever), `docs/USAGE.md` should call that out explicitly so operators know to put the gateway behind a reverse proxy that strips or rewrites CORS. Either approach works — the current default is the one that needs reconsidering before the OAuth path is exposed publicly.

nalyk:
**Memory-leak path in the Streamable HTTP session-init flow** — `src/transport.ts` ~lines 88-104:

```ts
transport = new StreamableHTTPServerTransport({
  sessionIdGenerator: () => randomUUID(),
  onsessioninitialized: (newSessionId: string) => {
    transports[newSessionId] = transport!;   // (1) registers in the map
  }
});

transport.onclose = () => {
  const sid = transport!.sessionId;
  if (sid) delete transports[sid];           // (2) cleans up — but only if onclose fires
};

await sessionServer.connect(transport);       // (3) if this throws → outer catch returns 500
```

Sequence under failure:

1. The SDK assigns a session ID and invokes `onsessioninitialized` → the entry is now in `transports[sid]`.
2. Something later inside `connect()` (or in `Server.connect`'s downstream wiring) throws — malformed initialize body, version mismatch, network setup failure, etc.
3. The outer `try/catch` around line ~121 emits a generic 500. It has no reference to `newSessionId`, so it can't clean up the map.
4. `transport.onclose` is not guaranteed to fire on a failed `connect()` — it depends on whether the SDK bound the lifecycle handlers before throwing.
5. Result: the map keeps a reference to a half-dead transport, which keeps the per-session `Server`, the `GitLabApi`, and the captured Bearer token alive for the lifetime of the process.

There's no concurrency cap and no idle reaper, so a caller who can repeatably trigger a `connect()` failure (e.g., POST `/mcp` with init payloads that pass the header check but fail SDK validation) can build zombie sessions until OOM.

Suggested fix — capture the session ID outside the callback and explicitly clean up on the catch path:

```ts
let initializedSid: string | undefined;
const newTransport = new StreamableHTTPServerTransport({
  sessionIdGenerator: () => randomUUID(),
  onsessioninitialized: (sid) => {
    initializedSid = sid;
    transports[sid] = newTransport;
  }
});

newTransport.onclose = () => {
  const sid = newTransport.sessionId;
  if (sid && transports[sid] === newTransport) {
    delete transports[sid];
  }
};

try {
  await sessionServer.connect(newTransport);
} catch (err) {
  if (initializedSid) delete transports[initializedSid];
  try { await newTransport.close?.(); } catch { /* best-effort */ }
  throw err;
}
transport = newTransport;
```

The `const newTransport` also lets you drop the two `transport!` non-null assertions in `onsessioninitialized` and `onclose` — TS can now narrow the closure capture without help.

Related (separate but same shape): the legacy SSE branch around line ~122 cleans up only via `req.on("close", ...)`, which fires on TCP close but not on SDK-initiated graceful close. Mirroring `transport.onclose` there with an idempotent cleanup helper would close the symmetrical gap. Worth folding into the same fix while you're in this file.

nalyk:
**Not a blocker — flagging for follow-up.** The `createMcpServer(token)` factory in `src/index.ts` (~line 711 to ~1841) wraps the original module-scope handler registration block, including the full ~1100-line `CallToolRequestSchema` switch over the 86 tools. The body of the wrapped block wasn't re-indented during the change — visually, the file still reads as if the handlers live at module scope.

Two consequences worth thinking about:

**1. Per-session allocation cost in OAuth mode.**
In PAT mode this runs once at startup — irrelevant. In OAuth mode the factory runs on every Streamable HTTP `POST /mcp` initialize and every legacy `GET /sse` connection. Each call re-allocates the `Server` instance, the `GitLabApi`, and (more meaningfully) two large closures over the entire 86-tool dispatch table. At N concurrent OAuth sessions you have N copies of those closures plus the GC pressure of releasing them when sessions end. The work is largely unnecessary, because the only per-session state is `gitlabApi` — the dispatch logic itself is identical across sessions.

**2. Readability and maintainability.**
Adding a new tool now requires editing inside a 1130-line function rather than appending a top-level `case`. The trailing `// end createMcpServer` comment at line ~1841 is the symptom — when a function needs a marker to find its closing brace, it's already too big. The unchanged indentation also makes future diffs noisier than necessary; any edit to a handler will look at first glance like an indented rewrite.

Suggested shape — extract the handlers to module-scope functions that take `gitlabApi` (and `readOnly`) as parameters, then the factory becomes a thin wiring layer:

```ts
// module scope — defined once, reused across sessions
async function handleListTools(readOnly: boolean): Promise<ListToolsResult> {
  return { tools: readOnly ? READ_ONLY_TOOLS : ALL_TOOLS };
}

async function handleCallTool(
  request: CallToolRequest,
  gitlabApi: GitLabApi
): Promise<CallToolResult> {
  switch (request.params.name) {
    case "list_projects":
      return formatProjects(await gitlabApi.listProjects(/* ... */));
    // ... rest of the switch, unchanged
  }
}

// factory shrinks to ~10 lines
export function createMcpServer(token: string): Server {
  const server = new Server(
    { name: "gitlab-mcp-server", version: packageJson.version },
    { capabilities: { tools: {} } }
  );
  const gitlabApi = new GitLabApi({ apiUrl: GITLAB_API_URL, token });

  server.setRequestHandler(ListToolsRequestSchema, () =>
    handleListTools(GITLAB_READ_ONLY_MODE));
  server.setRequestHandler(CallToolRequestSchema, (req) =>
    handleCallTool(req, gitlabApi));

  return server;
}
```

Per-session allocation drops to: one `Server`, one `GitLabApi`, two thin arrow closures. Adding a tool stays where it always was — at module scope in the dispatch table. The factory becomes inspectable at a glance.

Happy to take this on as a follow-up `refactor: extract MCP handlers from createMcpServer factory` PR after the current blockers land — no need to expand the scope of this one. Mainly wanted to flag the pattern before it calcifies and starts attracting "just one more case in the factory" PRs.

ecthelion77:
Hey @nalyk — all four must-fix items and the three technical suggestions are now addressed in a single squash commit rebased on latest `main`:

**Must-fix:**
1. ✅ CHANGELOG restored from `origin/main`, new work under `## [Unreleased]`
2. ✅ `package.json` → `0.4.0`
3. ✅ `chart/values.yaml` removed (belongs in #29)
4. ✅ 17 new tests covering Bearer extraction, Streamable HTTP lifecycle, session collision, CORS, and /healthz

**Technical suggestions implemented:**
- CORS: origin allowlist via `CORS_ALLOW_ORIGINS`; `*` only in PAT mode; OAuth mode denies by default. `Allow-Methods`/`Allow-Headers` only emitted when an origin is actually allowed.
- Memory leak: `initializedSid` capture + best-effort `close()` on `connect()` failure (StreamableHTTP). Idempotent `cleanupSse()` on both `req.close` and `transport.onclose` (SSE). Transport registered only after successful `connect()`.
- `AUTH_MODE` fail-fast: invalid values → immediate `process.exit(1)` with clear message.
- `/healthz`: returns `{"status":"ok","sessions":<n>}` or 503 when sessions exceed `HEALTHZ_MAX_SESSIONS`.
- `docs/OPERATIONS.md` added (fixes broken cross-refs).

**Not in this PR (follow-up):**
- `createMcpServer` handler extraction refactor (as you suggested)
- `/readyz` with GitLab reachability check (natural fit for #29)

Rebased on `c2057f9` (typescript 6.0.3 bump), build and all 58 tests pass. Ready for re-review.

nalyk:
@ecthelion77 — your response was excellent and the implementation is exactly what was asked for. I verified locally: the `initializedSid` cleanup pattern, idempotent `cleanupSse`, lazy CORS allowlist, AUTH_MODE fail-fast, `/healthz` with `HEALTHZ_MAX_SESSIONS` (which exceeds the original ask in a useful way), and 17 tests across all 5 categories — all match the proposals 1:1, and 58/58 vitest cases pass on Node 22.

There were a couple of mechanical regressions in the rebased squash commit: `package.json` reverted recent dependency bumps (typescript 6 from #35, express 5 from #36, the `vite ^8` security override from #38) and the `[0.3.2]` CHANGELOG entry that landed today during your response window was missing. Not your fault — main moved fast under your branch while you were turning around the review, mostly from PRs I merged in parallel. That's on me as the maintainer who landed those in your response window.

Rather than ask you to redo a rebase for shifts I caused, I rebased your branch onto current main, restored the dropped `package.json` deps and the `CHANGELOG` entry, regenerated the lockfile, and opened #42 with the result. Your commit and authorship are preserved — `git log` shows you as Author with me as Committer, and the squash-merge will carry a `Co-authored-by:` trailer so GitHub credits you on the contribution graph.

Closing this PR in favor of #42. Once #42's CI is green I'll squash-merge and ship 0.4.0 via npm Trusted Publishing.

Re your deferred items: the `createMcpServer` handler extraction makes sense as a standalone `refactor:` PR whenever you're up for it; `/readyz` with the GitLab reachability check is the right home in #29 once we land that one. No urgency on either.

Thanks for the discipline and for turning this around so cleanly.

nalyk:
Superseded by #42.

PR #56, opened 2026-05-05
ci(chart): add negative tests for fail-loud guards in validate job

### LINKED ISSUE #48: CI: add negative tests for chart fail-loud guards in validate job
## Context

The Helm chart ships **5 fail-loud guards** (PR #44):

1. Empty PAT in PAT mode without `existingSecret`
2. `podDisruptionBudget.minAvailable` >= `replicaCount`
3. Both `existingSecret` AND inline `secret.GITLAB_PERSONAL_ACCESS_TOKEN` set
4. Both `podDisruptionBudget.minAvailable` AND `.maxUnavailable` set
5. Invalid `AUTH_MODE` value

The `validate` job (`.github/workflows/build.yml:24–46`) runs `helm lint` and `helm template` with a single positive input set:

```
--set secret.GITLAB_PERSONAL_ACCESS_TOKEN=dummy-for-lint
```

## Problem

**Positive-only validation cannot detect guard regression.** If a refactor in `chart/templates/secret.yaml` accidentally drops the `fail "..."` call, `helm lint --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=dummy-for-lint` still passes (the dummy token satisfies the guard). The bug ships undetected, and the next operator who hits an empty token gets a silent misconfiguration instead of a clean error.

This is a known weakness — flagged in PR #44's "out of scope" follow-ups as *"`helm lint --set dummy-token` masks fail-loud guard regression"*.

## Acceptance criteria

- [ ] `validate` job adds a `negative-tests` step (or sibling job) that runs `helm template` with each of the 5 broken inputs and asserts each fails
- [ ] Each assertion checks **both** non-zero exit code AND that stderr contains the expected guard message (so silent failure or message regression is also caught)
- [ ] Skeleton:
  ```bash
  - name: Negative tests — fail-loud guards
    run: |
      set +e
      run_negative() {
        local name=$1; local expected=$2; shift 2
        OUT=$(helm template t chart/ "$@" 2>&1)
        RC=$?
        if [[ $RC -eq 0 ]]; then
          echo "::error::$name: expected failure, got success"; exit 1
        fi
        if ! grep -q "$expected" <<<"$OUT"; then
          echo "::error::$name: expected message '$expected' not in stderr"; exit 1
        fi
        echo "::notice::$name OK"
      }
      run_negative "empty-token" "GITLAB_PERSONAL_ACCESS_TOKEN" \
        --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=
      run_negative "pdb-deadlock" "minAvailable.*replicaCount" \
        --set podDisruptionBudget.enabled=true --set podDisruptionBudget.minAvailable=1 --set replicaCount=1 \
        --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=ok
      run_negative "secret-conflict" "existingSecret.*GITLAB_PERSONAL_ACCESS_TOKEN" \
        --set existingSecret=foo --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=ok
      run_negative "pdb-both-fields" "minAvailable.*maxUnavailable" \
        --set podDisruptionBudget.enabled=true --set podDisruptionBudget.minAvailable=1 --set podDisruptionBudget.maxUnavailable=1 \
        --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=ok
      run_negative "auth-mode" "AUTH_MODE" \
        --set config.AUTH_MODE=invalid \
        --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=ok
  ```
- [ ] Add a comment block in `chart/values.yaml` listing all 5 guards + their expected error message strings, so the negative tests stay aligned with template intent

## Files

- `.github/workflows/build.yml`
- `chart/values.yaml` — comment block documenting guards

## Priority

**P2 — CI quality.** Good first issue: well-scoped, no design tradeoffs, immediate value, non-blocking.

## References

- #44 — guards introduced (and noted as future-work)

cc @ecthelion77

ecthelion77:
## Summary

Adds a **"Negative tests — fail-loud guards"** step to the `validate` job in `.github/workflows/build.yml` that exercises all five chart guards with deliberately broken inputs. Each test asserts both a non-zero exit code and the expected error substring in the output.

## Guards covered

| # | Test name | Guard | Template / schema | Trigger |
|---|-----------|-------|-------------------|---------|
| 1 | `empty-pat-token` | Empty PAT in pat mode | `secret.yaml` | `--set secret.GITLAB_PERSONAL_ACCESS_TOKEN=""` |
| 2 | `pdb-deadlock` | minAvailable ≥ replicaCount | `pdb.yaml` | `minAvailable=1, replicaCount=1` |
| 3 | `secret-conflict` | Both existingSecret and inline PAT | `secret.yaml` | `existingSecret=foo` + `PAT=tok` |
| 4 | `pdb-both-fields` | Both minAvailable and maxUnavailable | `pdb.yaml` | Both set to 1 |
| 5 | `invalid-auth-mode` | Invalid AUTH_MODE | `values.schema.json` | `AUTH_MODE=invalid` |

## Changes

- `.github/workflows/build.yml` — new step after `Helm template smoke test`, before `Install helm-docs`
- `chart/values.yaml` — documentation comment block listing all guards + their error substrings

## Validation

- **CI tested on fork** — [run 25388453083](https://github.com/forterro/mcp-gitlab-server/actions/runs/25388453083) passed all steps including the new negative tests
- `helm-docs` regenerated: no README drift (verified with `helm-docs --check`)
- All 5 negative tests pass locally on Helm v3.17.3
- Positive `helm lint` / `helm template` still pass

## Notes

- The `run_negative` helper uses `set +e` to capture exit codes and `grep -qi` for case-insensitive substring matching
- Guard 2 requires `--set podDisruptionBudget.maxUnavailable=null` to clear the default from `values.yaml`
- Guard 5 is enforced by `values.schema.json` (enum constraint), not a template `fail` call

Closes #48

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,6 +45,65 @@ jobs:
             --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=dummy-for-lint \
             > /dev/null
 
+      - name: Negative tests — fail-loud guards
+        run: |
+          set +e
+          FAIL=0
+          run_negative() {
+            local name="$1"; local expected="$2"; shift 2
+            OUT=$(helm template t chart/ "$@" 2>&1)
+            RC=$?
+            if [[ $RC -eq 0 ]]; then
+              echo "::error::$name: expected failure, got success"
+              FAIL=1; return
+            fi
+            if ! grep -qi "$expected" <<< "$OUT"; then
+              echo "::error::$name: expected message '$expected' not in output"
+              echo "::error::$name: actual output: $OUT"
+              FAIL=1; return
+            fi
+            echo "::notice::$name OK"
+          }
+
+          # Guard 1: empty PAT in pat mode without existingSecret
+          run_negative "empty-pat-token" "GITLAB_PERSONAL_ACCESS_TOKEN is required when AUTH_MODE=pat" \
+            --set config.AUTH_MODE=pat \
+            --set config.HOST=127.0.0.1 \
+            --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=""
+
+          # Guard 2: PDB minAvailable >= replicaCount (deadlock)
+          run_negative "pdb-deadlock" "would deadlock node drains" \
+            --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=ok \
+            --set podDisruptionBudget.enabled=true \
+            --set podDisruptionBudget.minAvailable=1 \
+            --set podDisruptionBudget.maxUnavailable=null \
+            --set replicaCount=1
+
+          # Guard 3: both existingSecret and inline token
+          run_negative "secret-conflict" "Set EITHER existingSecret OR secret.GITLAB_PERSONAL_ACCESS_TOKEN" \
+            --set existingSecret=my-secret \
+            --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=tok
+
+          # Guard 4: PDB both minAvailable and maxUnavailable
+          run_negative "pdb-both-fields" "set EITHER minAvailable OR maxUnavailable" \
+            --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=ok \
+            --set podDisruptionBudget.enabled=true \
+            --set podDisruptionBudget.minAvailable=1 \
+            --set podDisruptionBudget.maxUnavailable=1
+
+          # Guard 5: invalid AUTH_MODE (rejected by values.schema.json)
+          run_negative "invalid-auth-mode" "values don't meet the specifications" \
+            --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=ok \
+            --set config.AUTH_MODE=invalid
+
+          # Guard 6: PAT mode on non-loopback host (GHSA-8jr5-6gvj-rfpf)
+          run_negative "pat-non-loopback" "AUTH_MODE=pat requires HOST" \
+            --set config.AUTH_MODE=pat \
+            --set config.HOST=0.0.0.0 \
+            --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=ok
+
+          if [[ $FAIL -ne 0 ]]; then exit 1; fi
+
       - name: Install helm-docs
         run: |
           curl -sL https://github.com/norwoodj/helm-docs/releases/download/v1.14.2/helm-docs_1.14.2_Linux_x86_64.tar.gz \
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-_Nothing yet. New entries land here between releases._
+### Added
+
+- **CI: negative tests for chart fail-loud guards (#48)** — the `validate`
+  job now runs `helm template` with six deliberately broken value sets and
+  asserts each guard fires (non-zero exit + expected error substring).
+  Includes the new `auth-validation.yaml` guard from #58 (GHSA-8jr5-6gvj-rfpf).
+- `chart/values.yaml` documentation comment block listing all six guards,
+  their source template, and the CI-matched error substring.
 
 ## [0.6.0] - 2026-05-05
 
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -1,5 +1,22 @@
 replicaCount: 1
 
+# ===========================================================================
+# Fail-loud guards
+# ===========================================================================
+# The chart templates include validation guards that abort with a clear error
+# message when values are misconfigured. CI runs negative tests to ensure
+# these guards are not accidentally removed during refactoring.
+#
+#   Guard                    Template              Expected error substring
+#   ~~~~~~~~~~~~~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#   Empty PAT in pat mode    secret.yaml           GITLAB_PERSONAL_ACCESS_TOKEN is required when AUTH_MODE=pat
+#   PDB deadlock             pdb.yaml              would deadlock node drains
+#   Secret conflict          secret.yaml           Set EITHER existingSecret OR secret.GITLAB_PERSONAL_ACCESS_TOKEN
+#   PDB both fields          pdb.yaml              set EITHER minAvailable OR maxUnavailable
+#   Invalid AUTH_MODE        schema.json           values don't meet the specifications
+#   PAT non-loopback host    auth-validation.yaml  AUTH_MODE=pat requires HOST
+# ===========================================================================
+
 image:
   repository: ghcr.io/yoda-digital/mcp-gitlab-server
   # -- Image tag. Overridden by CI at package time.


nalyk:
Hey @ecthelion77 — thanks for picking this up cleanly. The shape of `run_negative`, the `maxUnavailable=null` reset on guard 2, the accumulator that runs all five before failing, signed-off-by, fork CI link in the body — all the right calls.

Heads up: #58 (GHSA-8jr5-6gvj-rfpf) landed this morning under embargo — you couldn't have seen it coming. It changes the chart in ways that matter for these tests, so let me walk through what I'd want addressed before merge.

## What landed in #58

- A sixth fail-loud guard at `chart/templates/auth-validation.yaml` — refuses install when `AUTH_MODE=pat` is combined with non-loopback `HOST`. Chart-side counterpart to the runtime guard from the advisory.
- Two default flips in `chart/values.yaml`:
  - `AUTH_MODE`: `pat` → `oauth`
  - `HOST` is now explicit at `0.0.0.0` (chart needs it for Service routing)

## What needs to change

### 1. Test 1 (`empty-pat-token`) — goes red on rebase

`secret.yaml`'s guard predicate is `if and (eq AUTH_MODE "pat") (not PAT)`. Under the new `AUTH_MODE=oauth` default the `and` short-circuits — verified locally with helm 3.17 against current main, `helm template` returns 0 and the chart fully renders. The helper then trips on `expected failure, got success`.

`--set config.AUTH_MODE=pat` is enough to recover it. Pinning `HOST=127.0.0.1` too locks the test to this specific guard so it stays decoupled from Helm's template evaluation order — `auth-validation.yaml` is also active under PAT + non-loopback `HOST`, and even though `secret.yaml` currently wins when both fire, the explicit pin removes the dependency:

```yaml
run_negative "empty-pat-token" "GITLAB_PERSONAL_ACCESS_TOKEN is required when AUTH_MODE=pat" \
  --set config.AUTH_MODE=pat \
  --set config.HOST=127.0.0.1 \
  --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=""
```

### 2. Add a sixth test for `auth-validation.yaml`

Of the six guards now, this one matters most for security — it blocks the exact misconfiguration GHSA-8jr5-6gvj-rfpf was about. If a refactor drops it without anyone noticing, the same hole re-opens:

```yaml
run_negative "pat-non-loopback" "AUTH_MODE=pat requires HOST" \
  --set config.AUTH_MODE=pat \
  --set config.HOST=0.0.0.0 \
  --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=ok
```

### 3. Pin `expected` strings to fail-message text only

Three of the current substrings appear in normal happy-path rendered output too — `GITLAB_PERSONAL_ACCESS_TOKEN` is a literal Secret data-key, `minAvailable` is a YAML field on every PDB-enabled render. The RC gate saves them today; if the helper ever gets simplified the substring check turns into a tautology. Pinning to phrases unique to each `fail` (or to Helm's schema validator) costs nothing — these are verbatim against current main:

```yaml
run_negative "empty-pat-token"   "GITLAB_PERSONAL_ACCESS_TOKEN is required when AUTH_MODE=pat" ...
run_negative "pdb-deadlock"      "would deadlock node drains" ...
run_negative "secret-conflict"   "Set EITHER existingSecret OR secret.GITLAB_PERSONAL_ACCESS_TOKEN" ...
run_negative "pdb-both-fields"   "set EITHER minAvailable OR maxUnavailable" ...
run_negative "invalid-auth-mode" "config.AUTH_MODE must be one of" ...
```

The schema-validator string for test 5 also keeps it from matching `auth-validation.yaml` if the messages converge later.

### 4. Update the doc block in `chart/values.yaml` and the CHANGELOG entry

The values.yaml comment block lists 5 guards; current main has 6. The CHANGELOG entry in this PR also reads "five deliberately broken value sets" — same edit. The whole point of the comment block from #48 was "stay aligned with template intent."

## One nit worth doing now

`--set podDisruptionBudget.maxUnavailable=null` works today, but `null` semantics have wobbled across Helm releases. With the substring pinning above this becomes belt-and-braces — even if a future Helm changes how `null` parses, the test fails loud instead of silently matching the wrong guard's message.

(Two bash-hygiene asides — `set +e` re-arm and `2>&1` stream split — I'll open as a follow-up issue. Not for this PR.)

---

If you're up for it, I'd like to loop you into the threat-model discussion for the next iteration of this chart. The auth × bind matrix sits right on top of the transport work you wrote — you've got the context to spot what an external researcher would. Thanks again — this is good work.

ecthelion77:
@nalyk Thanks for the thorough review — all four points addressed:

1. **Test 1 fixed** — added `--set config.AUTH_MODE=pat --set config.HOST=127.0.0.1` so it stays decoupled from `auth-validation.yaml`'s evaluation.
2. **Test 6 added** — `pat-non-loopback` covering `auth-validation.yaml` (GHSA-8jr5-6gvj-rfpf guard).
3. **Expected strings pinned** to unique fail-message phrases — no overlap with happy-path rendered output.
4. **Doc block + CHANGELOG updated** — now reflects 6 guards.

Rebased on current `main` (v0.6.0 / #58 included). All 6 tests verified locally with Helm 3.17.

Re: threat-model discussion — happy to participate, just let me know where/when.

nalyk:
@ecthelion77 — fast turnaround, all four points landed. Test 5's the only one red on CI, and **that one's on me**: the substring I recommended (`config.AUTH_MODE must be one of`) is helm-version-fragile. I verified locally on Helm 3.17, but CI's `azure/setup-helm@v4` resolves to Helm 4.1.4, and the schema-validator format diverged between majors:

- Helm 3: `- config.AUTH_MODE: config.AUTH_MODE must be one of the following: "pat", "oauth"`
- Helm 4: `- at '/config/AUTH_MODE': value must be one of 'pat', 'oauth'`

The other 5 pass on both — re-checked locally with 3.17.4 and 4.1.4 against your branch.

Quickest fix: pin to Helm's own schema-error wrapper instead of the gojsonschema message it contains. The wrapper is identical on Helm 3 and Helm 4:

```yaml
run_negative "invalid-auth-mode" "values don't meet the specifications" \
  --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=ok \
  --set config.AUTH_MODE=invalid
```

Helm prepends that wrapper to every schema rejection. Template `fail` calls read `execution error at (template:line:col)` instead, so no cross-guard overlap. Update the matching row in the `chart/values.yaml` doc block too — either mirror the new substring or use a plain-English description like `AUTH_MODE not in [pat, oauth]`. Your call.

Sorry for the extra round. Lesson filed on my end: Helm 3 and Helm 4 reworded the gojsonschema output without touching the Helm wrapper, so a substring anchored on the inner message broke between versions while the wrapper text stayed stable. Once that one-liner lands and CI's green, ready to merge.

nalyk:
@ecthelion77 — 12/12 on both Helm 3.17.4 and 4.1.4 against `21cf575f`. Verified locally with both versions. The `actual output: $OUT` line you added on the substring-mismatch path is a nice touch — kind of thing that helps the next person debug a CI red without having to re-run anything locally.

Squash-merging now.

On the threat-model thread: I'll open it as a Discussion. Auth × bind matrix for `helm install` and `docker run` both, plus the substring-anchoring lesson from this round. Ping you when it's up.

ecthelion77:
Done — switched test 5 to `"values don't meet the specifications"` and updated the doc block. CI is green. Thanks for the diagnosis and the lesson on anchoring to wrapper text!

Issue #46, opened 2026-05-04
Split /healthz into /livez (always 200) and /readyz (real check) — current shape causes restart loop on session overflow


nalyk:
## Context

PR #28 (landed via #42) introduced `/healthz` with an `HEALTHZ_MAX_SESSIONS` 503 trigger (`src/transport.ts:121–131`). PR #29 (landed via #44) wires the Helm chart's **liveness AND readiness** probes to that same `/healthz` endpoint (`chart/values.yaml:80–87` — defaults `/healthz` for both, `chart/templates/deployment.yaml:73–87` — both probes share a configurable path).

## Problem

The current shape conflates **liveness** (is the process responsive?) with **readiness** (should this pod accept traffic?). At session overflow the failure mode is self-amplifying:

1. `transport.ts:124` — `/healthz` returns 503 when `sessionCount > HEALTHZ_MAX_SESSIONS`
2. kubelet's `livenessProbe` sees 503 → restarts the pod
3. All in-flight MCP sessions for that pod are dropped
4. Surviving pods absorb the released load → may themselves cross the threshold
5. Cascading restarts under peak load — exactly the wrong reaction

Liveness should answer *"is the JS event loop responsive?"* — which, at session overflow, is **yes**. Readiness should answer *"should this pod take more sessions?"* — which, at session overflow, is **no**. Conflated, they produce a self-amplifying outage.

## Proposed split

| Endpoint | Behavior | Wired to |
|---|---|---|
| `/livez` | Always 200 unless the HTTP server is unresponsive | `livenessProbe` |
| `/readyz` | 503 if `sessionCount > HEALTHZ_MAX_SESSIONS` (drop from service rotation, no restart) | `readinessProbe` |
| `/healthz` | Thin alias of `/readyz` for one minor cycle (deprecation), then remove | (legacy) |

## Acceptance criteria

- [ ] `src/transport.ts` exposes `/livez` (always 200, body `{"status":"ok"}`) and `/readyz` (current `/healthz` 503 logic moved here)
- [ ] `src/transport.test.ts` adds tests:
  - `/livez` returns 200 even when `transports` count exceeds `HEALTHZ_MAX_SESSIONS`
  - `/readyz` returns 503 in that case, 200 otherwise
- [ ] `chart/values.yaml`: `probes.liveness.path` defaults to `/livez`, `probes.readiness.path` defaults to `/readyz`
- [ ] `chart/templates/NOTES.txt` updated to reference both endpoints
[...]

PR #57, opened 2026-05-05
fix(docker): add HEALTHCHECK, COPY --chown, digest-pin, drop USE_SSE

### LINKED ISSUE #45: Dockerfile: add HEALTHCHECK, switch to COPY --chown, digest-pin base image, drop opinionated USE_SSE=true ENV
## Context

The Dockerfile shipped in #44 (commit 0c2af49) provides a solid security baseline: multi-stage build, non-root execution, dropped capabilities, read-only-rootfs compatible. Four hardening items deferred from the original review remain — all senior-grade hygiene, none of them user-visible bugs today, all harder to fix later.

### 1. No `HEALTHCHECK` directive
`Dockerfile` ships no `HEALTHCHECK`. Operators running the image with plain Docker (compose, swarm, Docker Desktop) get no liveness signal. Kubernetes uses its own probes so this is invisible there, but the image is intended to also work standalone.

### 2. `COPY` without `--chown=node:node`
Three `COPY` directives (lines 12, 17, 27, 31) land as root-owned, and the runtime stage finishes with `RUN chown -R node:node /app` (line 40). That re-ownership rewrites every file's metadata into a new layer. Modern Docker supports `COPY --chown` natively — same outcome, smaller image, fewer layers.

### 3. Base image not digest-pinned
`FROM node:24-alpine` (lines 7, 22) is pinned by tag, not by digest. `node:24-alpine` is a moving target — re-builds may pick up upstream base-image changes silently. For an auth-bearing MCP server, tag-only pinning is too loose. Dependabot's `docker` ecosystem keeps digests current automatically.

### 4. Opinionated `USE_SSE=true` baked into runtime ENV
Line 35 sets `USE_SSE=true`. The Helm chart already wires `USE_SSE` through ConfigMap (`chart/values.yaml` line 16), creating two places of truth. Stdio-only or Streamable-HTTP-only deployments inherit an SSE default they don't want — and `src/index.ts` already defaults to stdio when no transport flag is set.

## Acceptance criteria

- [ ] Add `HEALTHCHECK` (e.g. `HEALTHCHECK --interval=30s --timeout=3s CMD wget --quiet --tries=1 --spider http://127.0.0.1:${PORT:-3000}/livez || exit 1`). Use `/livez` once #2 lands; `/healthz` until then
- [ ] Switch all three `COPY` directives to `COPY --chown=node:node …` and remove the trailing `RUN chown -R …` layer
- [ ] Pin both `FROM node:24-alpine` lines by digest: `FROM node:24-alpine@sha256:<digest> AS builder` / runtime
- [ ] Add `docker` ecosystem to `.github/dependabot.yml` so digest pins stay current
- [ ] Helm chart: add `image.digest` to `chart/values.schema.json`, wire through `chart/templates/deployment.yaml` (use `repository@digest` when digest is set, fall back to `repository:tag`), and reflect in `chart/values.yaml`. Schema/template/values must land together — accepting a key the chart doesn't honor is worse than rejecting it. Locked-down schema (#50) means the chart `helm lint` will fail until all three are aligned, which is the intended forcing function
- [ ] Drop `USE_SSE=true` from runtime ENV; Helm chart's `config.USE_SSE` becomes the single source of truth
- [ ] `hadolint Dockerfile` clean
- [ ] PR #44's verification matrix re-runs green: local `docker build`, `docker run` + `curl /healthz` (or `/livez`), `npm test`

## Files

- `Dockerfile`
- `.github/dependabot.yml`
- `chart/values.schema.json`
- `chart/templates/deployment.yaml`
- `chart/values.yaml`

## Priority

**P2 — hardening.** None of the four are runtime bugs. Each is the kind of senior-grade hygiene gap that's trivial today and painful later (especially the digest-pin once images are widely consumed).

## References

- #44 — Dockerfile origin
- #50 / #53 — Helm chart schema lockdown (forces digest field to be added in this PR)
- Hadolint best-practice DL3007 (latest tag)
- BuildKit `COPY --chown` docs
- Sigstore + supply-chain integrity rationale for digest pinning

cc @ecthelion77

ecthelion77:
## Summary

Dockerfile hardening — four changes addressing all items in #45:

1. **HEALTHCHECK directive** — `wget --spider http://127.0.0.1:${PORT}/healthz` with 30s interval. Provides liveness for Docker/Compose/Swarm; Kubernetes continues to use its own probes.

2. **`COPY --chown=node:node`** — replaces `RUN chown -R node:node /app` layer. Same security outcome, smaller image (fewer layers, no metadata rewrite).

3. **Digest-pinned base image** — both `FROM node:24-alpine` stages use `@sha256:d1b3b4da...`. Dependabot `docker` ecosystem (already in `.github/dependabot.yml`) keeps the pin current automatically.

4. **Dropped `USE_SSE=true` ENV** — the chart's `config.USE_SSE` is now the sole source of truth. Stdio-only or Streamable-HTTP deployments no longer inherit an unwanted SSE default.

## Chart changes

- `image.digest` field added to `values.yaml`, `values.schema.json`, and `deployment.yaml` template
- When `image.digest` is set, the deployment uses `repository@digest` (precedence over tag)
- `chart/README.md` regenerated via `helm-docs` — no drift

## Validation

- **CI tested on fork** — [run 25389503089](https://github.com/forterro/mcp-gitlab-server/actions/runs/25389503089) passed: hadolint ✅, helm lint ✅, helm template ✅, negative tests ✅, helm-docs drift check ✅, docker build+push ✅
- `hadolint Dockerfile` clean (no warnings)
- `docker build` succeeds locally
- `helm template` works in both tag mode and digest mode

## Notes

- Uses `/healthz` for HEALTHCHECK (per issue text: use `/livez` once #46 lands)
- `additionalProperties: false` on the `image` object in schema means the new `digest` field must be declared — done
- `dependabot.yml` already had the `docker` ecosystem configured from a prior commit

Closes #45

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to `@yoda.digital/gitlab-mcp-server` are documented here.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.6.1] - Unreleased
 
 ### Added
 
@@ -15,6 +15,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Includes the new `auth-validation.yaml` guard from #58 (GHSA-8jr5-6gvj-rfpf).
 - `chart/values.yaml` documentation comment block listing all six guards,
   their source template, and the CI-matched error substring.
+- **Dockerfile: `HEALTHCHECK` directive** — plain Docker / Compose / Swarm
+  deployments now get built-in liveness via `wget /healthz`.
+- **Helm chart: `image.digest` support** — `values.yaml` + schema +
+  deployment template accept an optional `image.digest` field that takes
+  precedence over `image.tag` when set.
+
+### Changed
+
+- **Dockerfile: digest-pin base image** — both `FROM node:24-alpine` stages
+  now use `@sha256:…` digest pinning. Dependabot `docker` ecosystem (already
+  configured) keeps the pin current automatically.
+- **Dockerfile: `COPY --chown=node:node`** — replaces the `RUN chown -R`
+  layer with native BuildKit ownership. `USER node` is set before `npm ci`
+  so `node_modules/` are owned by `node:node` by construction.
+- **Dockerfile: `HEALTHCHECK --start-period` bumped to 10s** — accommodates
+  cold-start on constrained pods (`resources.requests.cpu: 50m`).
 
 ## [0.6.0] - 2026-05-05
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 # ---------------------------------------------------------------------------
 
 # -- Stage 1: build ----------------------------------------------------------
-FROM node:24-alpine AS builder
+FROM node:24-alpine@sha256:d1b3b4da11eefd5941e7f0b9cf17783fc99d9c6fc34884a665f40a06dbdfc94f AS builder
 
 WORKDIR /app
 
@@ -19,16 +19,21 @@ RUN npm run build
 
 
 # -- Stage 2: runtime --------------------------------------------------------
-FROM node:24-alpine
+FROM node:24-alpine@sha256:d1b3b4da11eefd5941e7f0b9cf17783fc99d9c6fc34884a665f40a06dbdfc94f
 
 WORKDIR /app
 
-# Install only production dependencies in runtime image
-COPY package.json package-lock.json ./
+# Run as non-root using built-in node user (uid 1000).
+# USER set before npm ci so node_modules are owned by node:node by construction.
+RUN chown node:node /app
+USER node
+
+# Install only production dependencies (runs as node — owned by construction)
+COPY --chown=node:node package.json package-lock.json ./
 RUN npm ci --omit=dev --ignore-scripts
 
 # Copy compiled server
-COPY --from=builder /app/dist ./dist
+COPY --chown=node:node --from=builder /app/dist ./dist
 
 ENV NODE_ENV=production \
     PORT=3000 \
@@ -36,8 +41,7 @@ ENV NODE_ENV=production \
 
 EXPOSE 3000
 
-# Run as non-root using built-in node user (uid 1000)
-RUN chown -R node:node /app
-USER node
+HEALTHCHECK --interval=30s --timeout=3s --start-period=10s --retries=3 \
+  CMD wget --quiet --tries=1 --spider http://127.0.0.1:${PORT:-3000}/healthz || exit 1
 
 ENTRYPOINT ["node", "dist/index.js"]
--- a/chart/README.md
+++ b/chart/README.md
@@ -41,6 +41,13 @@ The chart includes render-time validation:
 - **Empty PAT token** — fails if `AUTH_MODE=pat` with no token and no `existingSecret`.
 - **PDB deadlock** — fails if `minAvailable >= replicaCount` (would block node drains).
 - **Mutual exclusion** — fails if both `existingSecret` and inline `secret.GITLAB_PERSONAL_ACCESS_TOKEN` are set.
+- **PAT non-loopback** — fails if `AUTH_MODE=pat` with a non-loopback `HOST` (GHSA-8jr5-6gvj-rfpf).
+
+## Image pinning
+
+When `image.digest` is set (e.g. `sha256:abc123...`), the deployment uses
+`repository@digest` and the `image.tag` value is ignored. This allows
+deterministic rollouts without mutable tags. If both are set, digest wins.
 
 ## Multi-replica considerations
 
@@ -63,6 +70,7 @@ ingress controller with cookie-based routing). See
 | extraEnv | list | `[]` | Extra env vars (list of {name, value} or {name, valueFrom}) |
 | extraEnvFrom | list | `[]` | Extra envFrom (list of secretRef/configMapRef) |
 | fullnameOverride | string | `""` |  |
+| image.digest | string | `""` | Image digest (sha256:...). When set, takes precedence over tag. |
 | image.pullPolicy | string | `"IfNotPresent"` |  |
 | image.repository | string | `"ghcr.io/yoda-digital/mcp-gitlab-server"` |  |
 | image.tag | string | `"latest"` | Image tag. Overridden by CI at package time. Default "latest" is safe for local helm install; tagged releases use semver. |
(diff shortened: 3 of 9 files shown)


nalyk:
Hey @ecthelion77 — thanks for putting this together. The HEALTHCHECK shape (loopback inside container, `--start-period`, retries), `COPY --chown` over `RUN chown -R`, digest-pinning paired with the existing Dependabot docker ecosystem, the `image.digest` chart support with the precedence-over-tag template — all sound instincts. The smaller-image claim holds up, by the way: built both locally and PR 57's image is 282 MB vs current main at 325 MB. Real, not theoretical.

Heads up: like #56, this opened just before #58 (GHSA-8jr5-6gvj-rfpf) merged — same embargo timing. PR 58 flipped runtime defaults in a way that interacts with PR 57's removal of `USE_SSE=true`. Verified locally by building from your branch and running. Walking through it.

## The big one: default `docker run` is broken

Built `pr57-test:local` from `fix/dockerfile-hardening` and ran without env overrides:

```
$ docker run -d --name pr57-defaults -p 13000:3000 \
    -e GITLAB_PERSONAL_ACCESS_TOKEN=dummy pr57-test:local
$ docker logs pr57-defaults
GitLab MCP Server running with stdio transport(s) (auth: pat)
$ docker ps -a --filter name=pr57-defaults --format '{{.Status}}'
Exited (0) 5 seconds ago
$ docker inspect --format '{{.State.Health.Status}}' pr57-defaults
unhealthy
```

The chain: `ENV USE_SSE=true` removed → runtime defaults to stdio (`USE_SSE=false`, `USE_STREAMABLE_HTTP=false`, see `src/index.ts:152-153`) → stdio without a TTY hits EOF immediately → container exits cleanly with code 0 → HEALTHCHECK can't probe a dead container.

The image's whole shape — `EXPOSE 3000`, the HEALTHCHECK you're adding, the chart targeting `:3000/healthz` for liveness — is HTTP-oriented. Stdio in Docker without `-it` is non-functional. Three options, in order of how I'd take them:

**A. Keep an HTTP default to match the chart.** One ENV line. Undoes the most contentious of the four PR claims, but the chart already has `config.USE_SSE: "true"` (chart/values.yaml:21) so Docker stays consistent with Helm:

```dockerfile
ENV NODE_ENV=production \
    PORT=3000 \
    USE_SSE=true
```

**B. Move project-wide to Streamable HTTP.** Larger refactor — Dockerfile flips to `USE_STREAMABLE_HTTP=true` AND chart values flip to match. Doesn't have to happen here, but it's the right modernization at some point.

**C. Drop HEALTHCHECK + EXPOSE.** Image becomes a transport-agnostic wrapper. Cleanest philosophically, but loses what made HEALTHCHECK worth adding in the first place.

I'd take A in this PR and open a separate "modernize transport defaults" PR for B if we want it. Up to you.

## node_modules ownership regression

Verified inside the built image:

```
drwxr-xr-x    1 root root  /app
drwxr-xr-x   97 root root  /app/node_modules    ← was node:node before
```

`COPY --chown=node:node` only covers explicit COPY targets. The `node_modules/` tree from `RUN npm ci --omit=dev --ignore-scripts` runs as root and stays root-owned. The original `RUN chown -R node:node /app` caught this; removing it without replacement leaves `node:node` only partially preserved. `readOnlyRootFilesystem: true` in the chart blocks runtime writes, so it's not a security hole today — but the PR description's "Same security outcome" doesn't match the actual file ownership.

Two clean fixes — same layer count as today:

```dockerfile
# Option A: chown in the same layer as npm ci
RUN npm ci --omit=dev --ignore-scripts && chown -R node:node /app/node_modules

# Option B: USER node before npm ci (cleanest — node-owned by construction)
RUN chown node:node /app
USER node
COPY --chown=node:node package.json package-lock.json ./
RUN npm ci --omit=dev --ignore-scripts
COPY --chown=node:node --from=builder /app/dist ./dist
```

B is the senior version (npm ci runs as the unprivileged user). Either works.

## Version bump

`package.json` is at `0.6.0` (#58 cut that). Per CLAUDE.md P0, your changes need an in-PR bump to `0.6.1` to trigger the auto-publish workflow on merge.

## CHANGELOG categorization

The "drop USE_SSE=true" entry depends on which option from "the big one" lands. If **A**: the entry stays under `### Changed` (PR 57's intent is preserved, no consumer break). If **B** or **C**: move to `### Breaking changes` with migration text along the lines of:

> **Dockerfile: `USE_SSE=true` ENV removed.** Direct image consumers (`docker run`) now need `-e USE_SSE=true` (or `-e USE_STREAMABLE_HTTP=true`) explicitly, plus `-e HOST=0.0.0.0 -e AUTH_MODE=oauth` for network exposure — see the auth × bind matrix in `SECURITY.md`.

## Two nits while you're in there

- `HEALTHCHECK --start-period=5s` is tight for cold-start on a constrained pod (`resources.requests.cpu: 50m`). 10s is safer at no steady-state cost.
- Tag-vs-digest precedence is in the inline `values.yaml` comment and schema description, but not in `chart/README.md` prose. One sentence so an operator setting both doesn't wonder why their `:v0.7.0` tag bump isn't showing up.

---

The auth × bind matrix angle from #56 has a Docker counterpart that you'd be the right person to think through — "what does `docker run` see" sits right next to "what does `helm install` see", and #58 just shifted both. If you're up for it on either, ping me. Thanks for the depth — keep them coming.

ecthelion77:
@nalyk All points addressed — rebased on v0.6.0 and force-pushed:

**The big one (option A):** `USE_SSE=true` retained in ENV. Consistent with chart defaults, `docker run` produces a working HTTP server out of the box.

**node_modules ownership (option B):** `USER node` is now set *before* `npm ci`, so `node_modules/` are owned by `node:node` by construction. No orphaned root files.

**Version bump:** 0.6.0 → 0.6.1 in package.json + package-lock.json.

**CHANGELOG:** stays under `### Changed` (no consumer break since USE_SSE is kept). Removed the "drop USE_SSE" entry.

**Nits:**
- `--start-period` bumped to 10s
- Added "Image pinning" prose section in chart/README.md.gotmpl explaining digest-over-tag precedence
- Also added the PAT non-loopback guard to the fail-loud guards list in README

Re: auth × bind matrix for Docker — happy to think that through. LMK.

ecthelion77:
Rebased on upstream/main (now includes #56) — CHANGELOG conflict resolved. The 0.6.1 section now has both the CI negative-tests entries from #56 and the Dockerfile hardening entries from this PR. Ready for re-review.

nalyk:
@ecthelion77 — verified end-to-end and it's clean. Built the image, ran with the default `docker run` invocation (just `-e GITLAB_PERSONAL_ACCESS_TOKEN=dummy`), container stayed up and HEALTHCHECK transitioned to healthy. HTTP listener on 127.0.0.1:3000 in the logs, exactly the shape we wanted. The Big-One regression from last round is gone.

- `node_modules/` owned `node:node` inside the image — option B, `npm ci` runs as the unprivileged user, no orphan root-owned files.
- All 6 chart negative tests pass on Helm 3.17.4 and 4.1.4.
- `image.digest` renders `repository@digest` when set, falls back to `:tag` otherwise. Schema rejects garbage digests at lint time.
- helm-docs drift: zero. The `README.md.gotmpl` update keeps regenerations aligned — didn't think to ask for it, glad you did.

Squash-merging now. I'll open the auth × bind / Docker matrix Discussion and ping you on it.

PR #60, opened 2026-05-06
feat: split /healthz into /livez and /readyz probes

### LINKED ISSUE #46: Split /healthz into /livez (always 200) and /readyz (real check) — current shape causes restart loop on session overflow

ecthelion77:
Fixes #46

## Summary

Splits the existing `/healthz` endpoint into Kubernetes-idiomatic probes:

- **`/livez`** — Always returns `200 {"status":"ok"}`. Pure liveness: the event loop is alive. Docker `HEALTHCHECK` updated to use this endpoint.
- **`/readyz`** — Returns 503 when sessions exceed `HEALTHZ_MAX_SESSIONS`. Signals the pod should be removed from service rotation.
- **`/healthz`** — Retained as deprecated alias of `/readyz` for backward compatibility. Will be removed in 0.7.0.

## Changes

- `src/transport.ts`: Add `/livez` handler, rename existing logic to serve both `/readyz` and `/healthz`.
- `src/transport.test.ts`: Six new tests covering the three endpoints under normal and overloaded conditions.
- `chart/values.yaml`: `probes.liveness.path` → `/livez`, `probes.readiness.path` → `/readyz`.
- `chart/README.md`: Regenerated via helm-docs.
- `Dockerfile`: `HEALTHCHECK` uses `/livez` (liveness semantics).
- `docs/OPERATIONS.md`: Expanded health check docs with table, per-endpoint details, deprecation notice.
- `CHANGELOG.md`: Entries under 0.6.1 Unreleased.

## Behavior change

`>= maxSessions` now triggers 503 (was `>`). An operator who set `HEALTHZ_MAX_SESSIONS=100` previously got 200 at exactly 100 sessions; now they get 503. This matches "at capacity = not ready" semantics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Helm chart: `image.digest` support** — `values.yaml` + schema +
   deployment template accept an optional `image.digest` field that takes
   precedence over `image.tag` when set.
+- **`/livez` endpoint** — always returns `200 {"status":"ok"}` regardless of
+  session count. Intended for Kubernetes `livenessProbe`.
+- **`/readyz` endpoint** — returns 503 when session count exceeds
+  `HEALTHZ_MAX_SESSIONS`. Intended for Kubernetes `readinessProbe`.
 
 ### Changed
 
@@ -31,6 +35,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   so `node_modules/` are owned by `node:node` by construction.
 - **Dockerfile: `HEALTHCHECK --start-period` bumped to 10s** — accommodates
   cold-start on constrained pods (`resources.requests.cpu: 50m`).
+- **Helm chart probe defaults** — `probes.liveness.path` now defaults to
+  `/livez`, `probes.readiness.path` to `/readyz`.
+
+### Deprecated
+
+- **`/healthz`** — retained as alias of `/readyz` for backward compatibility.
+  Will be removed in 0.7.0. Note: the alias inherits the new `>=` threshold
+  semantic from `/readyz` (was `>` in 0.6.0). An operator at exactly
+  `HEALTHZ_MAX_SESSIONS` sessions now sees 503 where 0.6.0 returned 200.
 
 ## [0.6.0] - 2026-05-05
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,6 +42,6 @@ ENV NODE_ENV=production \
 EXPOSE 3000
 
 HEALTHCHECK --interval=30s --timeout=3s --start-period=10s --retries=3 \
-  CMD wget --quiet --tries=1 --spider http://127.0.0.1:${PORT:-3000}/healthz || exit 1
+  CMD wget --quiet --tries=1 --spider http://127.0.0.1:${PORT:-3000}/livez || exit 1
 
 ENTRYPOINT ["node", "dist/index.js"]
--- a/chart/README.md
+++ b/chart/README.md
@@ -79,7 +79,7 @@ ingress controller with cookie-based routing). See
 | podAnnotations | object | `{}` |  |
 | podDisruptionBudget | object | `{"enabled":false,"maxUnavailable":1}` | PodDisruptionBudget (recommended when replicaCount > 1). NOTE: SSE/Streamable HTTP transports hold in-memory session state. Multi-replica requires sticky sessions — see docs/OPERATIONS.md. |
 | podLabels | object | `{}` |  |
-| probes | object | `{"liveness":{"enabled":true,"initialDelaySeconds":5,"path":"/healthz","periodSeconds":30},"readiness":{"enabled":true,"initialDelaySeconds":3,"path":"/healthz","periodSeconds":10}}` | Liveness / readiness probes |
+| probes | object | `{"liveness":{"enabled":true,"initialDelaySeconds":5,"path":"/livez","periodSeconds":30},"readiness":{"enabled":true,"initialDelaySeconds":3,"path":"/readyz","periodSeconds":10}}` | Liveness / readiness probes |
 | replicaCount | int | `1` |  |
 | resources.limits.cpu | string | `"500m"` |  |
 | resources.limits.memory | string | `"256Mi"` |  |
--- a/chart/templates/NOTES.txt
+++ b/chart/templates/NOTES.txt
@@ -6,3 +6,7 @@
 
 2. In-cluster URL for MCP clients:
    http://{{ include "gitlab-mcp.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.service.port }}/sse
+
+3. Health probes:
+   Liveness:  /livez
+   Readiness: /readyz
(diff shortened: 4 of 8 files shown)


nalyk:
@ecthelion77 — clean PR. The cross-PR catch on the Dockerfile HEALTHCHECK (`/livez` instead of `/healthz`) was preemptive — that's the kind of thing reviewers usually have to flag. The deprecated-alias shape on `/healthz` is exactly right (path keeps working, marked for 0.7.0 removal, no operator break), and the test set covers the boundaries that matter — `/livez` stays 200 under forced overflow, `/readyz` flips to 503, `/healthz` matches `/readyz` bit-for-bit.

Verified locally:

- 34/34 vitest cases pass on this branch (6 new endpoint tests + 28 existing)
- All 6 chart negative tests from #56 still pass on Helm 3.17.4 and 4.1.4 — probe-path change didn't regress
- Built the image, ran with default `docker run`, HEALTHCHECK transitions to `healthy`, all three endpoints return as designed in normal and `HEALTHZ_MAX_SESSIONS=0` overflow conditions
- helm-docs drift: zero

Two small must-fix items before merge.

### 1. Inline deprecation comment for /healthz (#46 AC item)

#46 explicitly required "deprecation notice in CHANGELOG.md AND inline code comment." CHANGELOG has it; `src/transport.ts:276` doesn't. One line:

```ts
// /healthz is a deprecated alias of /readyz; removed in 0.7.0.
if (req.method === 'GET' && (pathname === '/readyz' || pathname === '/healthz')) {
```

So the deprecation is visible in the file, not just in CHANGELOG.

### 2. Surface the `> → >=` change in the /healthz deprecation entry

The PR body flags the threshold shift (was `>`, now `>=`). The CHANGELOG entry under `### Added /readyz` describes the new semantic; the entry under `### Deprecated /healthz` doesn't mention that the alias inherited it. Operators on /healthz won't see the boundary shift unless it's called out in the deprecated entry too.

Add to the `/healthz` deprecated entry:

> Note: the alias also picks up the new `>=` threshold semantic from `/readyz` (was `>` in 0.6.0). An operator at exactly `HEALTHZ_MAX_SESSIONS` sessions now sees 503 where 0.6.0 returned 200.

The semantic change itself is defensible — "MAX = at capacity = not ready" is the right read of the variable name. The ask is just to document the behavior change where /healthz operators will see it.

## One nit (not a blocker)

`chart/templates/NOTES.txt` is the post-install courtesy that prints the access URL. It doesn't currently mention any health endpoint, so AC item #4 in #46 ("update NOTES.txt to reference both endpoints") was technically moot — but a 2-line "Liveness probe: /livez · Readiness probe: /readyz" hint after the existing access lines would close that AC cleanly. Skip if you'd rather not touch it.

---

Threat-model thread (#59) is open whenever you have cycles. Once those two items land, this is good to merge.

ecthelion77:
Both must-fix items addressed + the NOTES.txt nit:

1. **Inline deprecation comment** — added above the `/readyz || /healthz` conditional in `src/transport.ts:276`.
2. **Threshold shift in deprecated entry** — CHANGELOG now notes the `> → >=` semantic change in the `/healthz` deprecated bullet.
3. **NOTES.txt** — added probe endpoint hints after the access URL section.

CI should go green shortly.

nalyk:
@ecthelion77 — verified end-to-end and clean. All three asks landed; you tightened the threshold-note phrasing and the NOTES.txt formatting beyond my drafts.

Verified locally on `6b78efc9`:

- 34/34 vitest cases pass
- All 6 chart negative tests from #56 pass on Helm 3.17.4 and 4.1.4
- Built the image, ran with default `docker run`, HEALTHCHECK transitions to `healthy` within 15s, all three endpoints return as designed
- Inline `/healthz` deprecation comment is at `src/transport.ts:276`, threshold-shift note is in the Deprecated entry, NOTES.txt has the probe hints
- helm-docs drift: zero

Squash-merging now. Discussion #59 is open whenever you have cycles — auth × bind matrix, Streamable HTTP modernization, and the substring-anchoring discipline are all there.

PR #97, opened 2026-05-25
fix(chart): allow 'enabled' and 'global' in values schema

ecthelion77:
## Problem

When `gitlab-mcp` is used as a sub-chart dependency in an umbrella chart, Helm injects two additional top-level properties into the sub-chart's values:

- `enabled` — from the `condition:` field in the parent's `Chart.yaml`
- `global` — inherited global values from the parent chart

Since `values.schema.json` has `"additionalProperties": false` at the root level, `helm template` fails with:

```
at '': additional properties 'enabled', 'global' not allowed
```

This blocks any ArgoCD or Flux deployment using this chart as a dependency.

## Fix

Add both properties to the schema's root `properties`:

- `enabled`: `{"type": "boolean"}`
- `global`: `{"type": "object", "additionalProperties": true}`

This is a standard pattern for Helm sub-charts with strict schemas (see [Helm docs on subcharts and global values](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/)).

## Testing

```bash
helm template test chart/ --set secret.GITLAB_PERSONAL_ACCESS_TOKEN=ok --set enabled=true --set global.imageRegistry=foo
```

Passes without error after the fix.


ecthelion77:
@nalyk Friendly ping — this fix is now blocking our upgrade path from 0.5.1 (forked) to 0.9.1 (upstream). Created issue #104 with full context.

The PR is a 17-line additive change with zero behavioral impact — just the two properties Helm requires for sub-chart usage (`enabled`, `global`). Would you be able to review/merge this soon? Thanks!

PR #99, opened 2026-05-27
feat: pipeline investigation tools (closes #64, supersedes #86)

### LINKED ISSUE #64: feat: improve pipeline job investigation with higher-level tools

nalyk:
## Path B reincarnation of #86

This PR supersedes #86 with the maintainer's round-3 follow-up fixes applied directly. Per [CONTRIBUTING.md § "Maintainer rebase pattern (Path B)"](../blob/main/CONTRIBUTING.md), the contributor's authorship on the original four commits is preserved via `git cherry-pick -x`. The remaining contract gaps identified in the round-3 review are landed as three maintainer commits on top of the contributor's work.

Closes #64. Supersedes #86 (preserved as the design+iteration audit trail).

## Contributor credit (4 commits, authorship preserved)

@ecthelion77 / Olivier Gintrand `<olivier.gintrand@forterro.com>`:

- `feat: add pipeline investigation tools (closes #64)` - initial implementation of `get_pipeline_summary`, `get_job_log_smart`, `list_pipeline_jobs + include_log_tail` extension per the agreed design from #64.
- `refactor: address Gemini Code Assist review feedback` - pulled `cleanLog()` out as a helper after the first automated review pass.
- `fix: address PR #86 review - bounded fan-out, error surfacing, pagination safety` - round-2: 11/12 items closed-as-described, plus unprompted helper extraction (`analyzeFailurePattern`, `deriveStageStatus`) and exported types (`FailurePattern`, `PipelineSummaryResponse`, `JobLogSmartResponse`).
- `fix: address PR #86 round-3 residuals - type safety, contract clarity` - round-3: 4 of 5 contract-clarity items closed-as-described.

The convergent three-reviewer audit (code-quality + silent-failure + type-design) moved the type-design rating from 2/5 → 4/5 → 4.6/5 across the three rounds. This is genuinely high-craft work; the reincarnation is for the maintainer to close two residual gaps that didn't make the contributor's last push, not a rewrite of the contribution.

## Maintainer round-3 follow-up (3 commits)

**`fix(pipeline-tools): close round-3 review residuals`** (`src/index.ts`, `src/gitlab-api.ts`):
1. **`list_pipeline_jobs (include_log_tail=true)` wrapper consistency.** The unified `{jobs, log_fetch_errors?}` response shape now applies on every path, including the previously-overlooked zero-failed-jobs fallback where `formatJobsResponse(jobs)` reverted the response to the legacy flat-array shape. This was the only PARTIAL item from the round-3 verification.
2. **`FailurePattern.shared_reason` gains `unreasoned_count`.** Convergent MEDIUM finding from the round-3 audit: when some failed jobs lack `failure_reason` (filtered out by `.filter(Boolean)`), the previous `count` understated the diagnosis scope without signal. `unreasoned_count = failedJobs.length - reasons.length` makes the gap explicit; `count === unreasoned_count + reasons.length` always.
3. **`FailurePattern.single.reason` normalization.** Empty-string `failure_reason` from GitLab now collapses to `null` via `||` rather than passing through `??` verbatim. Aligns with the `.filter(Boolean)` semantics in the multi-job branches.
4. **`get_job_log_smart.line_count` empty-log faithfulness.** `''.split('\n').length === 1` is a JS quirk; the response now reports `0` when `log === ''` (a path that became reachable in round 3 when `section`-not-found started returning empty log).

**`test(e2e): tighten pipelines assertions for round-3 contract fixes`** (`e2e/src/tests/pipelines.e2e.ts`):
- Replaces the single `expect(text.length).toBeGreaterThan(0)` smoke for `list_pipeline_jobs + include_log_tail` with two structured tests that parse the JSON wrapper and assert the contract on BOTH paths (failed-jobs-exist and zero-failed-jobs-fallback). The latter is a regression test for the fix above.

**`docs(changelog): record round-3 follow-up fixes for 0.9.0`** (`CHANGELOG.md`):
- Updates the `failure_pattern` entry to list all five variants and document `unreasoned_count`.
- Adds three Fixed entries for the round-3 follow-up items.

## Deferred to 0.9.1 (explicit in the round-3 review comment)

These were named as deferred at https://github.com/yoda-digital/mcp-gitlab-server/pull/86#issuecomment-4534477857 and are NOT addressed here:
- `truncated=true` false-positive at the exact 5000-job pagination boundary
- `page`/`per_page` still bare `z.number()` on `ListPipelineJobsSchema` (other numeric fields use `.int().min().max()`)
- `max_failed_jobs_with_logs: 0` as a second-channel disable path alongside `include_logs: false`
- Pure type-design aesthetic refactors (`StageStatus = Exclude<JobStatus, 'created'>`, `StageStatus` z.enum companion)

## Test coverage

- Unit tests: 86/86 pass (`npm test`)
- Coverage gate: PASS (88/88 tools have at least one test)
- TypeScript: clean (`npm run build`)
- E2E: triggered via `[e2e]` flag on the test commit; runs against a real GitLab CE in CI

## Merge plan

Per `feedback-pr-merge-strategy-preserve-vs-squash`: this is a multi-commit substantive PR - use `--merge` to preserve the audit trail (4 contributor commits + 3 maintainer commits on main). Squash would discard the contributor authorship + round-by-round history that makes this PR legible six months from now.

After merge: 0.9.0 release per the release-driven publish pattern.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,4 @@ implementation/
 mempalace.yaml
 .mempalace/
 entities.json
+.helm-pkg/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-_Nothing yet. New entries land here between releases._
+### Added
+
+- **`get_pipeline_summary` tool** — single-call pipeline investigation returning
+  pipeline details, jobs grouped by stage, and log tails for failed jobs. Includes
+  failure pattern detection and `max_failed_jobs_with_logs` cap (default: 5) to
+  prevent context blowup. (#64)
+- **`get_job_log_smart` tool** — intelligent log post-processor that strips ANSI
+  codes, GitLab timestamps, and section markers. Supports `tail`/`head`, section
+  extraction, and `error_only` filtering. (#64)
+- **`list_pipeline_jobs` extension** — new optional `include_log_tail` and
+  `log_tail_lines` parameters. When `include_log_tail=true`, failed jobs include
+  their cleaned log tail directly in the response. (#64)
 
 ## [0.8.1] - 2026-05-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `shared_reason`. Previously the mixed branch silently dropped failed jobs
   whose `failure_reason` was missing/empty from the visible reason histogram.
   (#99 codex)
+- **`get_job_log_smart` section extraction is now exact-match.** The previous
+  regex `section_start:\d+:NAME[^\n]*\n?` accepted a prefix match - requesting
+  `section: "build"` would silently return `build_extra` content with
+  `section_matched: true`. Now uses a lookahead `(?=[\r\n\[]|$)` after the
+  name to require GitLab's actual section delimiters. (#99 codex P2 round-2)
 
 ### Performance
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,43 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **`get_job_log_smart` tool** — intelligent log post-processor that strips ANSI
   codes, GitLab timestamps, and section markers. Supports `tail`/`head`, section
   extraction, and `error_only` filtering. (#64)
-- **`list_pipeline_jobs` extension** — new optional `include_log_tail` and
-  `log_tail_lines` parameters. When `include_log_tail=true`, failed jobs include
-  their cleaned log tail directly in the response. (#64)
+- **`list_pipeline_jobs` extension** — new optional `include_log_tail`,
+  `log_tail_lines`, and `max_log_tail_jobs` parameters. When `include_log_tail=true`,
+  failed jobs include their cleaned log tail directly in the response. (#64)
+- **`truncated` field** in `get_pipeline_summary` response — signals when pagination
+  cap (50 pages) was hit and job list may be incomplete.
+- **`section_matched` and `error_lines_matched` fields** in `get_job_log_smart`
+  response — explicit feedback on filter effectiveness.
+- **`log_fetch_errors` field** in pipeline summary — surfaces per-job log fetch
+  failures instead of silently swallowing them.
+
+### Changed
+
+- **Renamed** `log_lines` → `log_tail_lines` in `get_pipeline_summary` schema for
+  consistency with `list_pipeline_jobs`. (#64 review)
+- **`failure_pattern`** is now a discriminated union (`shared_reason | mixed |
+  no_failures | null`) instead of a plain string. (#64 review)
+- **Stage status derivation** now mirrors GitLab's aggregation: `allow_failure`
+  jobs no longer poison the stage; mixed success+skipped collapses to success.
+  (#64 review)
+- **Pagination** no longer relies on `X-Total` header (absent in GitLab EE);
+  uses `items.length < per_page` with a MAX_PAGES=50 safety cap. (#64 review)
+
+### Fixed
+
+- **`error_only` mode** in `get_job_log_smart` now correctly returns an empty log
+  (instead of the full log) when no error-like lines are found. (#64 review)
+- **Silent log fetch failures** — `getJobLogTails` and `getPipelineSummary` now
+  track and report errors per job instead of empty catch blocks. (#64 review)
+- **E2E tests** — assertions moved outside try/catch so test failures propagate
+  correctly; try/catch narrowed to the fetch step only. (#64 review)
+
+### Security
+
+- Schema parameters now have strict `.int().min().max()` bounds to prevent
+  abuse (e.g., `log_tail_lines` capped at 200, `max_log_tail_jobs` at 20).
+- XOR validation via `.refine()` on mutually exclusive parameters
+  (`pipeline_id`/`ref`, `tail`/`head`).
 
 ## [0.8.1] - 2026-05-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `section: "build"` would silently return `build_extra` content with
   `section_matched: true`. Now uses a lookahead `(?=[\r\n\[]|$)` after the
   name to require GitLab's actual section delimiters. (#99 codex P2 round-2)
+- **`stripSections` now consumes both CR and LF after the marker.** GitLab's
+  section line `section_*:NNN:name\r\x1B[0K\n` collapses to
+  `section_*:NNN:name\r\n` after ANSI stripping; the previous regex tail
+  `[\r\n]?` consumed only one of CR/LF, leaving an orphan `\n` per marker.
+  Cleaned logs had spurious blank lines and `tail: N` could shift by an
+  empty line. New regex tail `\r?\n?` consumes the full CRLF combo.
+  (#99 codex P2 round-3)
 
 ### Performance
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,8 +30,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Renamed** `log_lines` → `log_tail_lines` in `get_pipeline_summary` schema for
   consistency with `list_pipeline_jobs`. (#64 review)
-- **`failure_pattern`** is now a discriminated union (`shared_reason | mixed |
-  no_failures | null`) instead of a plain string. (#64 review)
+- **`failure_pattern`** is now an exhaustive discriminated union
+  (`no_failures | single | shared_reason | mixed | unknown`) instead of a
+  plain string. The `shared_reason` variant carries `unreasoned_count` so
+  callers can tell when some failed jobs lacked a `failure_reason` (vs.
+  every failure sharing the same reason). (#64 review)
 - **Stage status derivation** now mirrors GitLab's aggregation: `allow_failure`
   jobs no longer poison the stage; mixed success+skipped collapses to success.
   (#64 review)
@@ -46,6 +49,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   track and report errors per job instead of empty catch blocks. (#64 review)
 - **E2E tests** — assertions moved outside try/catch so test failures propagate
   correctly; try/catch narrowed to the fetch step only. (#64 review)
+- **`list_pipeline_jobs` (`include_log_tail=true`)** now always emits the
+  unified `{jobs, log_fetch_errors?}` wrapper, including the zero-failed-jobs
+  fallback path that previously reverted to the legacy flat-array shape.
+  (#64 round-3 follow-up)
+- **`get_job_log_smart`** returns `line_count: 0` for an empty log instead of
+  the JS-quirk `1` from `''.split('\n')`. Faithful counts on the
+  section-not-found and `error_only` no-match paths. (#64 round-3 follow-up)
+- **`FailurePattern.single.reason`** normalizes empty-string `failure_reason`
+  to `null` (was passed through verbatim by `??`). Aligns with `.filter(Boolean)`
+  semantics used by the multi-job variants. (#64 round-3 follow-up)
 
 ### Security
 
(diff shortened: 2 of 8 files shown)


nalyk:
Both bot reviews addressed in three follow-up commits on this branch.

## @chatgpt-codex-connector P2 (`mixed.unreasoned_count`)

Valid finding - the `mixed` branch was the asymmetric sibling. When `shared_reason` got `unreasoned_count` in the round-3 follow-up, I missed extending the same field to `mixed`, leaving `sum(reasons) < failedJobs.length` silently possible for the multi-distinct-reason case.

Fixed in `d9b1a2e` (`fix(pipeline-tools): mixed.unreasoned_count for parity with shared_reason`). The type union now reads:

```ts
| { kind: 'mixed'; reasons: Record<string, number>; unreasoned_count: number }
```

Invariant `sum(reasons) + unreasoned_count === failedJobs.length` holds across both multi-job variants.

## @gemini-code-assist 5x MEDIUM (perf)

All five flagged sites were the same shape: `log.split('\n').slice(-N).join('\n')` (or its filter equivalent) allocating O(N) substrings just to retain O(K). For multi-MB GitLab CI logs that's 10K+ small object allocations per fetch. Real concern.

Fixed in `1145db8` (`perf(pipeline-tools): O(K) log tail/head/count and non-allocating section extraction`). Three new private helpers on `GitLabApi`:

- `logTail(s, n)` - walks `\n` from the end via `lastIndexOf`; O(n) memory.
- `logHead(s, n)` - mirror via `indexOf` from the start.
- `countLines(s)` - single-pass `charCodeAt` loop; no substring allocation.

Updated all 5 callsites flagged:
1. `getPipelineSummary` log tail → `this.logTail(this.cleanLog(rawLog), logLines)`
2. `getJobLogTails` log tail → same helper
3. `getJobLogSmart` section extraction → `sectionEnd.lastIndex = startIdx; sectionEnd.exec(rawLog)` (no `rawLog.slice`)
4. `getJobLogSmart` error_only → single `/^.*(?:error|fatal|failed|exception|traceback|panic).*$/gim` regex match
5. `getJobLogSmart` tail/head + line_count → uses the new helpers

Behavior preserved (86/86 unit tests still pass); `''.split('\n').length === 1` quirk handling stays correct via `countLines` returning 0 for empty input.

CHANGELOG entry updated in `044d6d0` under a new `### Performance` section + cross-references to the codex finding.

Thanks to both bots for the catches.

---

@codex review

nalyk:
Codex P2 (section prefix-match) addressed in commit `a48e00c`. The regex was indeed too permissive:

```ts
// before: silently matched build_extra when section: "build" was requested
section_start:\d+:${escapedName}[^\n]*\n?

// after: pin the name to a GitLab section delimiter via lookahead
section_start:\d+:${escapedName}(?=[\r\n\[]|$)[^\n]*\n?
```

The lookahead requires CR, LF, `[option=...]`, or EOS immediately after the name - which is exactly what GitLab's section marker grammar emits, never `_anything`.

Coverage: 4 new unit tests in `src/gitlab-api.test.ts` lock the contract:
- Prefix rejection (the original bug case)
- Exact name followed by CR
- Exact name with `[collapsed=true]` marker
- Exact name picked over a sibling prefix-shared section

90/90 unit tests pass. CHANGELOG updated.

Thanks for the catch on the re-review pass.

---

@codex review

nalyk:
Codex P2 round-3 (`stripSections` CRLF consumption) addressed in commit `734b26b`.

The regex tail was `[\r\n]?` - matches AT MOST one of `\r` or `\n`. GitLab's actual section line is `section_*:NNN:name\r\x1B[0K\n`, which after `stripAnsi` collapses to `section_*:NNN:name\r\n`. The old pattern consumed only `\r`, leaving every `\n` orphan in the cleaned log. Result: spurious blank lines, and `tail: N` could shift by N empty lines worst case.

Fix:

```ts
// before:
section_(start|end):\d+:[^\r\n]*[\r\n]?

// after:
section_(start|end):\d+:[^\r\n]*\r?\n?
```

`\r?\n?` consumes any CRLF combination (CR alone, LF alone, CR+LF, or empty) after the section name. No orphan `\n` left behind.

Coverage: 2 new unit tests in `src/gitlab-api.test.ts`:
- Structural assertion: cleaned log contains no `\n\n` sequence where markers used to live
- `tail: 2` over a section_end-terminated log returns the last real line, not an orphan blank

93/93 unit tests pass. CHANGELOG updated under `### Fixed`.

Thanks again for the catch. This is the third round you've improved this PR, and each finding has been a real bug.

---

@codex review

nalyk:
Pre-push audit pass closed (codex R4 + 9 new findings from in-house agents). Following challenge that the bots were catching real bugs my own review process kept missing, I dispatched `pr-review-toolkit:silent-failure-hunter` and `code-simplifier:code-simplifier` on the maintainer-side diff before this push. They surfaced 9 actionable issues the prior 3-agent + 4 codex + gemini rounds had missed.

**Fixed in commit `6bb0352`:**

HIGH (silent fallbacks):
- `pipeline_id: 0` no longer silently fetches the latest pipeline. Schema is now `z.number().int().positive().optional()` and the runtime branch uses `!== undefined`.
- `log_fetch_capped: { fetched, total_failed }` field added to both `get_pipeline_summary.summary` and the `list_pipeline_jobs + include_log_tail` wrapper. Previously when failed jobs exceeded the per-tool cap (`max_failed_jobs_with_logs` default 5, `max_log_tail_jobs` default 10), trailing failures were silently dropped from log-fetching with no signal.

MEDIUM:
- `section: ""` now rejected at parse via `z.string().min(1)`. Previously the truthy check silently returned the full log with `section_matched: null`.

Codex P2 R4 (sections_found collapsed-marker leak):
- The discoverability loop now strips the `[collapsed=true]` suffix from captured names. Previously surfaced `"script[collapsed=true]"` which a caller could not pass back as `section:` (end markers carry no `[option]`).

Code-simplifier:
- Section markers are now stripped regardless of `strip_ansi`. Previously a caller setting `strip_ansi: false` to inspect raw ANSI output would see raw `section_start:NNN:name\r\x1B[0K` bytes leak into tail/head windows.
- `analyzeFailurePattern` precomputes `unreasoned_count` once instead of recomputing in each branch.
- `stripSections` docstring now states the strip-ANSI-first precondition explicitly.

LOW:
- Dropped the `i` flag on section regexes. GitLab section names are identifiers; case-insensitive matching would create silent ambiguity between sibling sections like `Build` vs `build`. Aligns with the "exact-match" claim from the R2 fix.

**Coverage:** +7 new unit tests (6 schema rejection assertions + 1 strip_ansi-false regression). 101/101 unit tests pass. CHANGELOG entries in commit `29f1a81`.

**Deferred to 0.9.1** (documentation/edge-case, not silent fallbacks):
- `include_log_tail` with non-failed `scope` is a no-op (doc fix only)
- `truncated` field semantics clarification (doc)
- `error_only` keyword list is English-only (doc)

@codex review

nalyk:
Codex R5 (2 P2) + pre-push silent-failure-hunter (1 MEDIUM + 1 LOW) closed in commit `fd46700`.

**Codex R5-1 (P2)**: `logTail`/`countLines` treated trailing `\n` as a new empty line. `tail: 1` on `"ERROR\n"` returned `""`; `log_tail_lines: 50` returned only 49 lines. Fix uses an `endIdx` bound (NOT a `slice(0, -1)` copy) so the O(tail) memory contract stays intact.

**Codex R5-2 (P2)**: `stripSections` regex `\r?\n?` didn't consume `\x1B[0K` between CR and LF, leaving orphan clear-control fragments when `strip_ansi: false`. Extended tail to `\r?(?:\x1B\[[0-9;]*[a-zA-Z])*\n?` - matches inline ANSI between CR and LF, works in both pre-stripped and raw flows.

**Pre-push silent-failure B1 (MEDIUM)**: the R5-1 fix made truncated/non-truncated branches return different shapes (`"L"` vs `"L\n"`). Closed with an unconditional single-`\n` trim at the end of `getJobLogSmart` so the `log` field is shape-stable.

**Pre-push silent-failure B2 (LOW)**: the initial R5-1 patch allocated `log.slice(0, -1)` on every call - O(N) regression vs the documented O(tail) promise. Refactored to use `endIdx` bound; zero extra allocation.

Coverage: +3 new unit tests (CRLF clear-control strip, single-line tail, 50-line tail). 104/104 unit tests pass. CHANGELOG entries in commit `b1cda8a` under a new `### Fixed (round 5)` block.

PR #99 is now at 18 commits (4 contributor + 14 maintainer). Bot+agent findings closed in this PR: 5 codex P2 + 5 gemini perf MEDIUM + 11 pre-push agent (HIGH/MEDIUM/LOW). That's 21 real bugs surfaced and fixed through this review cycle - a sizeable share of which were latent in the original code, not introduced by review iterations.

Thanks for the relentlessness.

---

@codex review

